PR #48, opened 2016-02-28
SerialFirmata: store&forward incoming bytes

jnsbyr:
timed buffering of incoming bytes at lower baud rates typically result
in message transfers instead of single byte transfers to the host (host
offloading)

### DIFF
--- a/examples/ConfigurableFirmata/ConfigurableFirmata.ino
+++ b/examples/ConfigurableFirmata/ConfigurableFirmata.ino
@@ -178,6 +178,9 @@ OneWireFirmata oneWire;
 #include <StepperFirmata.h>
 StepperFirmata stepper;
 
+//uncomment SERIAL_STORE_AND_FORWARD to collect bytes received by serial port until the receive buffer 
+// gets filled or a data gap is detected to avoid forwarding single bytes at baud rates below 50000
+//#define SERIAL_STORE_AND_FORWARD
 #include <SerialFirmata.h>
 SerialFirmata serial;
 
--- a/examples/ConfigurableFirmata/ConfigurableFirmata.ino
+++ b/examples/ConfigurableFirmata/ConfigurableFirmata.ino
@@ -178,9 +178,6 @@ OneWireFirmata oneWire;
 #include <StepperFirmata.h>
 StepperFirmata stepper;
 
-//uncomment SERIAL_STORE_AND_FORWARD to collect bytes received by serial port until the receive buffer 
-// gets filled or a data gap is detected to avoid forwarding single bytes at baud rates below 50000
-//#define SERIAL_STORE_AND_FORWARD
 #include <SerialFirmata.h>
 SerialFirmata serial;
 
--- a/src/SerialFirmata.cpp
+++ b/src/SerialFirmata.cpp
@@ -55,10 +55,10 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
         {
           long baud = (long)argv[1] | ((long)argv[2] << 7) | ((long)argv[3] << 14);
           serial_pins pins;
-#if defined(SERIAL_STORE_AND_FORWARD)    
+#if defined(FIRMATA_SERIAL_PORT_RX_BUFFERING)    
           lastAvailableBytes[portId] = 0;
           lastReceive[portId] = 0;
-          maxCharDelay[portId] = 50000 / baud; // 8N1 = 10 bits per char -> 50 bits max. char/char delay 
+          maxCharDelay[portId] = 50000 / baud; // 8N1 = 10 bits per char, max. 50 bits -> 50000 = 50bits * 1000ms/s
 #endif              
           if (portId < 8) {
             serialPort = getPortFromId(portId);
@@ -287,7 +287,7 @@ void SerialFirmata::checkSerial()
 
   if (serialIndex > -1) {
 
-#if defined(SERIAL_STORE_AND_FORWARD)
+#if defined(FIRMATA_SERIAL_PORT_RX_BUFFERING)
     unsigned long currentMillis = millis();
 #endif
 
@@ -309,7 +309,7 @@ void SerialFirmata::checkSerial()
       if (availableBytes > 0) {
         bool read = true;
 
-#if defined(SERIAL_STORE_AND_FORWARD)
+#if defined(FIRMATA_SERIAL_PORT_RX_BUFFERING)
         // check if reading should be delayed to collect some bytes before
         // forwarding (for baud rates significantly below 57600 baud)
         if (maxCharDelay[portId]) {
@@ -334,7 +334,9 @@ void SerialFirmata::checkSerial()
             numBytesToRead = bytesToRead;
           }
           
+#if defined(FIRMATA_SERIAL_PORT_RX_BUFFERING)
           lastAvailableBytes[portId] -= numBytesToRead; 
+#endif
 
           // relay serial data to the serial device
           while (numBytesToRead > 0) {
--- a/src/SerialFirmata.cpp
+++ b/src/SerialFirmata.cpp
@@ -49,162 +49,159 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
     byte mode = argv[0] & SERIAL_MODE_MASK;
     byte portId = argv[0] & SERIAL_PORT_ID_MASK;
 
-    switch (mode) {
-      case SERIAL_CONFIG:
-        {
-          long baud = (long)argv[1] | ((long)argv[2] << 7) | ((long)argv[3] << 14);
-          serial_pins pins;
-
-          if (portId < 8) {
-            serialPort = getPortFromId(portId);
-            if (serialPort != NULL) {
-              pins = getSerialPinNumbers(portId);
-              if (pins.rx != 0 && pins.tx != 0) {
-                Firmata.setPinMode(pins.rx, PIN_MODE_SERIAL);
-                Firmata.setPinMode(pins.tx, PIN_MODE_SERIAL);
-                // Fixes an issue where some serial devices would not work properly with Arduino Due
-                // because all Arduino pins are set to OUTPUT by default in StandardFirmata.
-                pinMode(pins.rx, INPUT);
+    if (portId < SERIAL_READ_ARR_LEN)
+    {
+      switch (mode) {
+        case SERIAL_CONFIG:
+          {
+            long baud = (long)argv[1] | ((long)argv[2] << 7) | ((long)argv[3] << 14);
+            serial_pins pins;
+#if defined(SERIAL_STORE_AND_FORWARD)
+            lastAvailableBytes[portId] = 0;
+            lastReceive[portId] = 0;
+            maxCharDelay[portId] = 50000 / baud; // 8N1 = 10 bits per char -> 50 bits max. char/char delay
+#endif
+            if (portId < 8) {
+              serialPort = getPortFromId(portId);
+              if (serialPort != NULL) {
+                pins = getSerialPinNumbers(portId);
+                if (pins.rx != 0 && pins.tx != 0) {
+                  Firmata.setPinMode(pins.rx, PIN_MODE_SERIAL);
+                  Firmata.setPinMode(pins.tx, PIN_MODE_SERIAL);
+                  // Fixes an issue where some serial devices would not work properly with Arduino Due
+                  // because all Arduino pins are set to OUTPUT by default in StandardFirmata.
+                  pinMode(pins.rx, INPUT);
+                }
+                ((HardwareSerial*)serialPort)->begin(baud);
               }
-#if defined(SERIAL_STORE_AND_FORWARD)    
-              lastAvailableBytes[portId] = 0;
-              lastReceive[portId] = 0;
-              maxCharDelay[portId] = 50000 / baud; // 8N1 = 10 bits per char -> 50 bits max. char/char delay 
-#endif              
-              ((HardwareSerial*)serialPort)->begin(baud);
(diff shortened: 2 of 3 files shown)


soundanalogous:
Wouldn't moving the call to `serial.update()` in the main loop to a timed interval (an interval based on the baud rate) accomplish something similar? However I see the advantage in your approach could be that the delay is per port and looks at the number of bytes in the buffer.

src/SerialFirmata.cpp
```diff
@@ -296,27 +310,43 @@ void SerialFirmata::checkSerial()
...
-        } else {
-          numBytesToRead = bytesToRead;
+      int availableBytes = serialPort->available();
+      if (availableBytes > 0) {
+        bool read = true;
+        
+#if defined(SERIAL_STORE_AND_FORWARD)    
+        // check if reading should be delayed to collect some bytes before forwarding (for baud rates up to 19200)
+        if (maxCharDelay[portId]) {          
+          // inter character delay exceeded or more than 48 bytes available or more bytes available than required
```
soundanalogous: Why 48 bytes specifically?

src/SerialFirmata.cpp
```diff
@@ -296,27 +310,43 @@ void SerialFirmata::checkSerial()
...
-          numBytesToRead = bytesToRead;
+      int availableBytes = serialPort->available();
+      if (availableBytes > 0) {
+        bool read = true;
+        
+#if defined(SERIAL_STORE_AND_FORWARD)    
+        // check if reading should be delayed to collect some bytes before forwarding (for baud rates up to 19200)
+        if (maxCharDelay[portId]) {          
+          // inter character delay exceeded or more than 48 bytes available or more bytes available than required          
+          read = (lastAvailableBytes[portId] > 0 && (currentMillis - lastReceive[portId]) >= maxCharDelay[portId]) || (bytesToRead == 0 && availableBytes >= 48) || (bytesToRead > 0 && availableBytes >= bytesToRead);
```
soundanalogous: break this up into multi-line. Try not to exceed 100 chars / line if possible

src/SerialFirmata.cpp
```diff
@@ -66,6 +66,11 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
                 // because all Arduino pins are set to OUTPUT by default in StandardFirmata.
                 pinMode(pins.rx, INPUT);
               }
+#if defined(SERIAL_STORE_AND_FORWARD)
```
soundanalogous: Define this block at the top of the SERIAL_CONFIG case. That way you only need to declare it once, since the only variable it depends on is `portId`.

jnsbyr:
If only one serial port is used, timing the serial.update() may be enough. But if you use two ports even at the same baud rate the incoming data frequency will differ. If you wait to long you get an overrun. If you don't wait long enough, it will not help much. My approach has a little bit more management overhead but makes the best of every baud rate without requiring manual tuning.

src/SerialFirmata.cpp
jnsbyr: I expect hardware serial and software serial ports to have 64 bytes of receive buffer. Processing a Firmata loop requires a variable amount of time. I chose this value to stay clear of a buffer overrun but this is not completely foolproof.

src/SerialFirmata.cpp
jnsbyr: done, commit pending

src/SerialFirmata.cpp
jnsbyr: done, commit pending
Note: addressing a serial port above 15 will result in an array bounds violation, we could add a check in line 51.

src/SerialFirmata.cpp
```diff
@@ -296,27 +310,43 @@ void SerialFirmata::checkSerial()
...
-        if (bytesToRead == 0 || (serialPort->available() <= bytesToRead)) {
-          numBytesToRead = serialPort->available();
-        } else {
-          numBytesToRead = bytesToRead;
+      int availableBytes = serialPort->available();
+      if (availableBytes > 0) {
+        bool read = true;
+        
+#if defined(SERIAL_STORE_AND_FORWARD)    
+        // check if reading should be delayed to collect some bytes before forwarding (for baud rates up to 19200)
```
soundanalogous: Change text within parenthesis to "for rates up to 49999"  - that is if I'm understanding this correctly - or is it only good for 50000 / 2?

src/SerialFirmata.cpp
soundanalogous: yes that would be useful

src/SerialFirmata.cpp
soundanalogous: Makes sense. Could make this value configurable at some point, but hard coding at 48 for now seems fine.

src/SerialFirmata.cpp
jnsbyr: Well, it's an integer division. If the baud rate is higher than 50000 the division result is zero milliseconds and then a timed store&forward is no longer possible, because the timer resolution is only milliseconds. Because of ```(currentMillis - lastReceive[portId]) >= maxCharDelay[portId])``` and assuming the Firmata loop requiring more than 1 millisecond (depending on hardware) one can assume that the effect of the store&forward may cease earlier. Maybe "for baud rates significantly below 57600 baud" is a better description.

src/SerialFirmata.cpp
jnsbyr: added
```C
    if (portId < SERIAL_READ_ARR_LEN)
    {
...
    }
  }
  return false;
```
.. and SERIAL_READ_ARR_LEN is "only" 12, not even the expected 16, but it is the right choise and enough for 4 software serial ports

src/SerialFirmata.cpp
```diff
@@ -49,152 +49,159 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
...
-            serialPort = getPortFromId(portId);
-            if (serialPort != NULL) {
-              pins = getSerialPinNumbers(portId);
-              if (pins.rx != 0 && pins.tx != 0) {
-                Firmata.setPinMode(pins.rx, PIN_MODE_SERIAL);
-                Firmata.setPinMode(pins.tx, PIN_MODE_SERIAL);
-                // Fixes an issue where some serial devices would not work properly with Arduino Due
-                // because all Arduino pins are set to OUTPUT by default in StandardFirmata.
-                pinMode(pins.rx, INPUT);
+    if (portId < SERIAL_READ_ARR_LEN)
```
soundanalogous: change to:
```c++
if (portId >= SERIAL_READ_ARR_LEN) return false;
```

jnsbyr: done and committed

soundanalogous:
Cool! I'll test this with my local setup and if there are no surprises will merge. Eventually I'll want this to be user-definable (rather than conditionally compiled) so I'll add a command to the Serial feature definition once I figure out how I want users to specify this mode. That command could both enable the 'store & forward' mode as well as optionally pass in a threshold (currently set to 48 bytes) if helpful.

soundanalogous:
Had a chance to test your code. There are a couple of issues, documented inline.

examples/ConfigurableFirmata/ConfigurableFirmata.ino
```diff
@@ -178,6 +178,9 @@ OneWireFirmata oneWire;
 #include <StepperFirmata.h>
 StepperFirmata stepper;
 
+//uncomment SERIAL_STORE_AND_FORWARD to collect bytes received by serial port until the receive buffer 
+// gets filled or a data gap is detected to avoid forwarding single bytes at baud rates below 50000
+//#define SERIAL_STORE_AND_FORWARD
```
soundanalogous: With Arduino, libraries are compiled before the sketch file runs so setting a define here does nothing. It needs to be set in the .h file.

src/SerialFirmata.cpp
```diff
@@ -296,27 +305,45 @@ void SerialFirmata::checkSerial()
...
-        } else {
-          numBytesToRead = bytesToRead;
+#if defined(SERIAL_STORE_AND_FORWARD)
+        // check if reading should be delayed to collect some bytes before
+        // forwarding (for baud rates significantly below 57600 baud)
+        if (maxCharDelay[portId]) {
+          // inter character delay exceeded or more than 48 bytes available or more bytes available than required
+          read = (lastAvailableBytes[portId] > 0 && (currentMillis - lastReceive[portId]) >= maxCharDelay[portId])
+                 || (bytesToRead == 0 && availableBytes >= 48) || (bytesToRead > 0 && availableBytes >= bytesToRead);
+          if (availableBytes > lastAvailableBytes[portId]) {
```
soundanalogous: There is something wrong is the logic here, but I haven't had time to look into it far enough to figure out exactly what is wrong. The issue is that this conditional is only true a couple of times when my test application first runs. Therefore `read == true` forever after that.

My test is using an Arduino Mega with an Adafruit GPS shield (at 9600 baud) wired to RX1 on the Mega.

jnsbyr: Sorry, I missed merging a standalone line - see next commit.

jnsbyr:
Another short note about the other magical number involved: 50000. It stands for "50bits * 1000ms/s" - the scaling with 1000 is required because we use milliseconds. The "50bits" is the interesting part. This number should be short enough to separate two serial messages send in sequence. Often a much higher duration is possible and only in a few cases you might need a lower value. If a human types manually, the bytes will usually still be send one by one and that is the expected behaviour.

soundanalogous:
The GPS module I test with sends large messages - far bytes than can ever even be received in a single message so it's an interesting module to test serial with.

src/SerialFirmata.h
```diff
@@ -142,6 +146,12 @@ class SerialFirmata: public FirmataFeature
     int serialBytesToRead[SERIAL_READ_ARR_LEN];
     signed char serialIndex;
 
+#if defined(SERIAL_STORE_AND_FORWARD)    
+    unsigned long lastReceive[SERIAL_READ_ARR_LEN];
+    unsigned char maxCharDelay[SERIAL_READ_ARR_LEN];
+    int lastAvailableBytes[SERIAL_READ_ARR_LEN];
```
soundanalogous: this line needs to be declared outside if the ifdef because it is used outside of an ifdef on line 337 in SerialFirmata.cpp

src/SerialFirmata.h
```diff
@@ -23,6 +23,10 @@
 #include <SoftwareSerial.h>
 #endif
 
+//uncomment SERIAL_STORE_AND_FORWARD to collect bytes received by serial port until the receive buffer 
+// gets filled or a data gap is detected to avoid forwarding single bytes at baud rates below 50000
+//#define SERIAL_STORE_AND_FORWARD
```
soundanalogous: what do you think about renaming `SERIAL_STORE_AND_FORWARD` to `FIRMATA_SERIAL_BUFFER_INPUT`?

src/SerialFirmata.cpp
```diff
@@ -48,13 +48,18 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
...
     switch (mode) {
       case SERIAL_CONFIG:
         {
           long baud = (long)argv[1] | ((long)argv[2] << 7) | ((long)argv[3] << 14);
           serial_pins pins;
-
+#if defined(SERIAL_STORE_AND_FORWARD)    
+          lastAvailableBytes[portId] = 0;
+          lastReceive[portId] = 0;
+          maxCharDelay[portId] = 50000 / baud; // 8N1 = 10 bits per char -> 50 bits max. char/char delay
```
soundanalogous: change comment to: "8N1 = 10 bits per char -> 50 bits max. 50000 = 50bits * 1000ms/s"

jnsbyr: its about a delay, so if wie drop "char/char delay" I suggest:
// 8N1 = 10 bits per char, max. 50 bits -> 50000 = 50bits * 1000ms/s

soundanalogous: sgtm

src/SerialFirmata.h
jnsbyr: Thx, good to know that it also works in different setups.

To reduce the data size for variables if store&forward is not used I suggest to keep the define in the header file and add another around line 337.

src/SerialFirmata.h
jnsbyr: Both are not perfect. There is always a possible confusion with the host link. If you prefer the latter I'm in, but here is another suggestions: ```SERIAL_PORT_RX_BUFFERING```.

soundanalogous: That works as well, but prepend with `FIRMATA_`. I have been having issues with defines being overwritten by other libraries or even changes in the Arduino core since they are global. I have not been so good at following this myself but I need to get in the habit of it, or start using enums instead.

jnsbyr: done with ```FIRMATA_SERIAL_PORT_RX_BUFFERING```, commit pending

jnsbyr: Just had another idea: Firmata uses the term "serial pin" for its non-host serial ports. So maybe we should call it ```FIRMATA_SERIAL_RX_PIN_BUFFERING```?

soundanalogous: I think it's okay as is. I forgot about this earlier, but the better approach is to provide a way for the user to enable RX buffering through an extension to the [Serial feature protocol](https://github.com/firmata/protocol/blob/master/serial.md) at that point these ifdefs will not longer be necessary (the code will remain but will only run if a boolean is set).

jnsbyr: Good idea, this is even better than several ifdefs cluttering up the code.

PR #49, opened 2016-03-06
setup ability to enable rx buffering per serial port

soundanalogous:
Next step is to add a command to the protocol to enable the user to enable RX buffering.

### DIFF
--- a/src/SerialFirmata.cpp
+++ b/src/SerialFirmata.cpp
@@ -55,11 +55,15 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
         {
           long baud = (long)argv[1] | ((long)argv[2] << 7) | ((long)argv[3] << 14);
           serial_pins pins;
-#if defined(FIRMATA_SERIAL_PORT_RX_BUFFERING)
           lastAvailableBytes[portId] = 0;
           lastReceive[portId] = 0;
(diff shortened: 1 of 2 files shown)


soundanalogous:
@jnsbyr fyi

soundanalogous:
That's a good suggestion. It also opens up the possibility of allowing the user to specify the delay value as an optional parameter:

```
// command to enable RX Buffering for a specified port with option to override default
// char delay value
0  START_SYSEX         (0xF0)
1  SERIAL_MESSAGE      (0x60)
2  SERIAL_EXTEND       (0x00) // OR with portId (0x09 = SERIAL_EXTEND | SW_SERIAL1)
3  SERIAL_RX_BUFFERING (0x02) // up to 127 extended commands
4  charDelay           [optional] (bits 0 - 6)
5  charDelay           [optional] (bit 7) // may only need 1 byte if charDelay is limited to 127 max
6  END_SYSEX           (0xF7)
```

jnsbyr:
Maybe you can add some more (inline) documentation, e.g.:
```C
// char delay value [ms] to detect the end of a message, defaults to 50 bits * 1000 / baud rate
// a value of 0 will disable RX buffering, resulting in single byte transfers to the host with 
// baud rates below approximately 56k (varies with CPU speed)
```

PR #50, opened 2016-04-21
Fix IS_PIN_INTERRUPT in Boards.h

drjrkuhn:
Interrupt numbers start at zero. The function `digitalPinToInterrupt()` correctly returns interrupt number `0` for `pin 2`, but this makes `IS_PIN_INTERRUPT(2)` false. I found the bug because the `FirmataEncoder` library checks `IS_PIN_INTERRUPT()` to determine whether a digital pin can be used as an encoder. It was incorrectly reporting pin 2 on an Arduino as incapable of being used as an encoder.

The correct way is to check whether `digitalPinToInterrupt()` returns a valid interrupt other than `NOT_AN_INTERRUPT` (=`-1`).


soundanalogous:
As an alternative, what about adding the following just below the last board definition:

```c++
#if defined(digitalPinToInterrupt) && defined(NOT_AN_INTERRUPT)
#define IS_PIN_INTERRUPT(p)     (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)
#else
#define IS_PIN_INTERRUPT(p)     (0)
#endif
```

And then removing `IS_PIN_INTERRUPT` from the individual board definitions. The only conflict will be with the Sanguino definition but that does not concern me.

drjrkuhn:
Good point about checking whether NOT_AN_INTERRUPT is defined! However, the original digitalPinToInterrupt() check is already specialized throughout Boards.h, so distributing the fix makes a sort of sense.

I propose the following macro that mirrors the concept of digitalPinHasPWM.

```
#if defined(digitalPinToInterrupt) && defined(NOT_AN_INTERRUPT)
#define digitalPinHasInterrupt(p)     (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)
#else
#define digitalPinHasInterrupt(p)     (0)
#endif
```

Now the digitalPinToInterrupt(p) code throughout the individual boards definitions can be changed to digitalPinHasInterrupt(p)

I found the `(digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)` test in the ESP8266 section of the [arduino firmata Boards.h](https://github.com/firmata/arduino/blob/master/Boards.h) file. So the root firmata repository has the same issue.

*Unfortunately, I did the original pull request just by editing in github. I'll close this pull request and start a new one with the updated patch and some information on how I tested!*

soundanalogous:
But there is no need to define the macro for each board. A single definition in Boards.h will work for all current (and even future) board entries in Boards.h. Simplifies things. However I can make that change later if necessary.

drjrkuhn:
I am convinced. Luckily, I had not had a chance to upload the new pull request yet. 
I'm testing a single common IS_PIN_INTERRUPT() macro instead of the per-board definitions. 

```
#if defined(digitalPinToInterrupt) && defined(NOT_AN_INTERRUPT)
#define IS_PIN_INTERRUPT(p)     (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)
#else
#define IS_PIN_INTERRUPT(p)     (0)
#endif

```
I'll submit a new pull request if my FirmataEncoder test code reports the correct pin ENCODER modes on my uno, nano, and mega.

PR #51, opened 2016-04-22
Fix for IS_PIN_INTERRUPT

drjrkuhn:
#### Rationale
ConfigurableFirmata adds support for the [FirmataEncoder](https://github.com/firmata/FirmataEncoder) library.

FirmataEncoder uses the macro `IS_PIN_INTERRUPT()` found in ConfigurableFirmata's [Boards.h](https://github.com/firmata/ConfigurableFirmata/blob/master/src/utility/Boards.h) to determine and report which pins support interrupts and may thus support an [Encoder](https://www.pjrc.com/teensy/td_libs_Encoder.html). 

The old macro in turn called `digitalPinToInterrupt()` to determine interrupt capabilities. However, digitalPinToInterrupt() correctly reports interrupt numbers of zero for pin 2 and -1 `(NO_ENCODER_AVAILABLE)` for non-interrupt pins. As a result, the FirmataEncoder library labeled most pins *except* pin 2 as supporting an ENCODER.

#### Fix in pull request
The `IS_PIN_INTERRUPT` is now defined globally for all boards as
```c++
#if defined(digitalPinToInterrupt) && defined(NOT_AN_INTERRUPT)
#define IS_PIN_INTERRUPT(p)     (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)
#else
#define IS_PIN_INTERRUPT(p)     (0)
#endif
```

The macro `digitalPinToInterrupt()` is supplied by the Arduino library in `pins_arduino.h`. It should be correctly defined for any arduino/genuino variant. All individual board-level checks have been removed from Boards.h.

#### Testing
I found the bug while extending the [firmata4j java host library](https://github.com/kurbatov/firmata4j) to support the [FirmataEncoder](https://github.com/firmata/FirmataEncoder) protocol.
I have made the extension available in [a branch of firmata4j](https://github.com/drjrkuhn/firmata4j/tree/addEncoder) for testing. To test, compile and upload a ConfigurableFirmata library that enables [FirmataEncoder](https://github.com/firmata/FirmataEncoder) support. 

- Run the [Example.java](https://github.com/drjrkuhn/firmata4j/blob/addEncoder/src/main/java/org/firmata4j/Example.java)  test application found in the package `org.firmata4j`. 
- Right click on an encoder field in the GUI to attach or detach it to two pins.
- The Example app uses the [DeviceReport](https://github.com/drjrkuhn/firmata4j/blob/addEncoder/src/main/java/org/firmata4j/ui/DeviceReport.java) class to print a human readable report of the pin support modes on standard out when any pin or encoder is changed. 
- Tests may be run with FirmataEncoder support disabled in ConfigurableFirmata. Verify that no pin is reported as ENCODER capable.

#### Test results
I tested encoder support on an Arduino Uno, Nano, and Mega 2560. With this patch, all pins now correctly report their interrupt available status (aka ENCODER mode available) on these three boards.

### DIFF
--- a/src/utility/Boards.h
+++ b/src/utility/Boards.h
@@ -143,9 +143,9 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #endif
 
 #if defined(digitalPinToInterrupt) && defined(NOT_AN_INTERRUPT)
-#define digitalPinHasInterrupt(p)     (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)
+#define IS_PIN_INTERRUPT(p)     (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)
 #else
-#define digitalPinHasInterrupt(p)     (0)
+#define IS_PIN_INTERRUPT(p)     (0)
 #endif
 
 // Arduino Duemilanove, Diecimila, and NG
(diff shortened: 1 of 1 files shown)


drjrkuhn:
Sorry that this was split into two commits. I am still getting used to GitHub and pull-requests.

PR #57, opened 2016-07-22
@PlatformIO Library Registry manifest file

valeros:
Description
* This library in Web Registry: http://platformio.org/lib/show/363/ConfigurableFirmata
* Specification: [library.json](http://docs.platformio.org/en/latest/librarymanager/config.html)
* Library Manager: http://docs.platformio.org/en/latest/librarymanager/index.html

Requested by a user in platformio/platformio-libmirror#69

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,16 @@
+{
+  "name": "ConfigurableFirmata",
+  "keywords": "interface, protocol",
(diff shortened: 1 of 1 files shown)


soundanalogous:
So I assume going forward I just update the version in this file?

valeros:
Yep, just update the version.

PR #71, opened 2017-05-29
Initial pass at ESP32 support

soundanalogous:
Wi-Fi transport only for now.
Tested on Sparkfun ESP32 Thing board.

This is sorta working when the board is configured as a TCP server (default) or client. However establishing a connection currently requires some patience. I suspect there may be issues in the ESP32 WiFi library because board locks up while trying to establish a connection with the AP more than half the time. It could also be a power issue. I've found the connection to be much more stable when using a USB power supply vs using my laptop usb port, but even with a power supply I still have to hit reset on the board a few times. I suspect a power supply > 500 mA is required (something in the 1000 to 2000 mA range is probably best).

Currently the procedure is restart the board, wait a couple of seconds then start the Firmata client application. It may take several seconds for the Firmata client to connect. Also, if reading analog pins, there are very noticeable lags. I suspect a more robust buffering scheme is necessary for Firmata to support boards like this.

I have no immediate plans to enable using ESP32 boards with Firmata using a Serial/USB transport or a BLE transport.

Firmata client examples interfacing with the ESP32 via a TCP client or server connection can be found [here](https://github.com/soundanalogous/firmata-client-examples/commit/8a943c189957b16cc3ac1f789fdd91ca367c98cd) (using firmata.js).

This branch will not be merged until WiFi connectivity and buffering is improved (could use help on figuring out an approach for the later), so checkout the esp-thing branch in the mean time.

### DIFF
--- a/examples/ConfigurableFirmataWiFi/ConfigurableFirmataWiFi.ino
+++ b/examples/ConfigurableFirmataWiFi/ConfigurableFirmataWiFi.ino
@@ -14,7 +14,7 @@
   Copyright (C) 2013 Norbert Truchsess. All rights reserved.
   Copyright (C) 2014 Nicolas Panel. All rights reserved.
   Copyright (C) 2015-2016 Jesse Frush. All rights reserved.
-  Copyright (C) 2009-2016 Jeff Hoefs.  All rights reserved.
+  Copyright (C) 2009-2017 Jeff Hoefs.  All rights reserved.
   Copyright (C) 2016 Jens B. All rights reserved.
 
   This library is free software; you can redistribute it and/or
@@ -24,7 +24,7 @@
 
   See file LICENSE.txt for further informations on licensing terms.
 
-  Last updated by Jeff Hoefs: August 14th, 2016
+  Last updated by Jeff Hoefs: May 28th, 2017
 */
 
 /*
@@ -43,6 +43,7 @@
   - ESP8266 WiFi board compatible with ESP8266 Arduino core
   - Arduino WiFi Shield 101
   - Arduino WiFi Shield (or clone)
+  - ESP32 board compatible with ESP32 Arduino core
 
   Follow the instructions in the wifiConfig.h file (wifiConfig.h tab in Arduino IDE) to
   configure your particular hardware.
@@ -53,6 +54,7 @@
     https://github.com/arduino-libraries/WiFi101)
   - ESP8266 requires the Arduino ESP8266 core v2.1.0 or higher which can be obtained here:
     https://github.com/esp8266/Arduino
+  - ESP32 requires the Arduino ESP32 core: https://github.com/espressif/arduino-esp32
 
   In order to use the WiFi Shield 101 with Firmata you will need a board with at least 35k of Flash
   memory. This means you cannot use the WiFi Shield 101 with an Arduino Uno or any other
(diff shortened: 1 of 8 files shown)


src/AnalogInputFirmata.cpp
```diff
@@ -52,7 +52,8 @@ void AnalogInputFirmata::reportAnalog(byte analogPin, int value)
         // Send pin value immediately. This is helpful when connected via
         // ethernet, wi-fi or bluetooth so pin states can be known upon
         // reconnecting.
-        Firmata.sendAnalog(analogPin, analogRead(analogPin));
+        // Firmata.sendAnalog(analogPin, analogRead(analogPin));
+        Firmata.sendAnalog(analogPin, analogRead(analogInputToDigitalPin(analogPin)));
```
soundanalogous: This should really be fixed in the esp32 arduino library. analogRead should be able to accept the analog pin number or the digital pin number.

src/AnalogOutputFirmata.cpp
```diff
@@ -20,6 +20,11 @@
 #include "AnalogFirmata.h"
 #include "AnalogOutputFirmata.h"
 
+// Hack to compile for ESP32 boards until arduino-esp32 supports analogWrite
+#ifdef ESP32
+void analogWrite(byte pin, int value);
```
soundanalogous: Necessary because Arduino compiles all cpp files whether or not they are included.

src/ConfigurableFirmata.h
```diff
@@ -100,8 +100,8 @@
 #define SYSEX_SAMPLING_INTERVAL 0x7A // same as SAMPLING_INTERVAL
 
 // pin modes
-//#define INPUT                 0x00 // defined in Arduino.h
```
soundanalogous: the esp32 library redefines `INPUT`, `OUTPUT` and `ANALOG`

src/AnalogInputFirmata.cpp
```diff
@@ -79,7 +80,11 @@ void AnalogInputFirmata::handleCapability(byte pin)
 {
   if (IS_PIN_ANALOG(pin)) {
     Firmata.write(PIN_MODE_ANALOG);
+#ifdef DEFAULT_ADC_RESOLUTION
+    Firmata.write(DEFAULT_ADC_RESOLUTION);
```
soundanalogous: This should be added for all of the StandardFirmata variants in the [firmata/arduino repo](https://github.com/firmata/arduino) as well.

examples/ConfigurableFirmataWiFi/ConfigurableFirmataWiFi.ino
```diff
@@ -317,6 +343,9 @@ char wep_key[] = "your_wep_key";
 #elif defined(ESP8266_WIFI) && defined(SERIAL_DEBUG)
 #define IS_IGNORE_PIN(p)  ((p) == 1)
 
+#elif defined(ESP32_WIFI) && defined(SERIAL_DEBUG)
+#define IS_IGNORE_PIN(p)  ((p) == 1 || (p) == 3)
```
soundanalogous: I have not been able to determine if this is absolutely necessary for the ESP32 since the connection is so flaky.

examples/ConfigurableFirmataWiFi/ConfigurableFirmataWiFi.ino
```diff
@@ -339,12 +368,18 @@ DigitalOutputFirmata digitalOutput;
 #include <AnalogInputFirmata.h>
 AnalogInputFirmata analogInput;
 
+// analogWrite not supported for ESP32 boards
+#ifndef ESP32
```
soundanalogous: I hate to have to do this... will probably not merge this branch until `analogWrite` is added to the esp32 library.

pgrawehr:
Closing this, as ESP32 support has now been merged (with help of this PR as well). Works for me with both USB and Wifi connection. Thanks to https://github.com/espressif/arduino-esp32 the ESP32 can be programmed and used like an Arduino.

PR #77, opened 2018-01-08
fixed mask used for stepType

dylanbannon:
Version 2.6.0 of the Firmata protocol
https://github.com/firmata/protocol/blob/master/accelStepperFirmata.md
says that, during accelStepper configuration, bits 4-6 of the interface byte encode the step type. To get bits 4-6, we should AND the interface byte with a mask of 0x0e.

### DIFF
--- a/src/AccelStepperFirmata.cpp
+++ b/src/AccelStepperFirmata.cpp
@@ -97,7 +97,7 @@ boolean AccelStepperFirmata::handleSysex(byte command, byte argc, byte *argv)
       if (stepCommand == ACCELSTEPPER_CONFIG) {
         interface = argv[index++];
         wireCount = (interface & 0x70) >> 4; // upper 3 bits are the wire count
-        stepType = (interface & 0x07) >> 1; // bits 4-6 are the step type
+        stepType = (interface & 0x0e) >> 1; // next 3 bits are the step type
         stepOrMotorPin1 = argv[index++]; // Step pin for driver or MotorPin1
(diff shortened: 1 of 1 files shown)


src/AccelStepperFirmata.cpp
```diff
@@ -97,7 +97,7 @@ boolean AccelStepperFirmata::handleSysex(byte command, byte argc, byte *argv)
       if (stepCommand == ACCELSTEPPER_CONFIG) {
         interface = argv[index++];
         wireCount = (interface & 0x70) >> 4; // upper 3 bits are the wire count
-        stepType = (interface & 0x07) >> 1; // bits 4-6 are the step type
```
soundanalogous: The mask is right here but the comment is wrong. It should read: `// bits 1-3 are the step type`.

According to the protocol definition: 
Bit 0 is for the enable pin. 
Bits 1-3 indicate the step type. 
Bits 4-6 indicate the wire count.

soundanalogous: Actually you're right, the mask should be 0x0e for bits 1-3. Both the mask and the comment were wrong.

soundanalogous:
LGTM

src/AccelStepperFirmata.cpp
dylanbannon: Haha, cool. I was worried for a second there that I had no clue what I was talking about, since I am definitely not a computer scientist

PR #80, opened 2018-05-16
Implements EXTENDED_ANALOG_READ message

mdlima:
As discussed on https://github.com/firmata/protocol/issues/103, this implements the EXTENDED_ANALOG_READ message:

```
// The client must request an analog read, then the client will respond
0  START_SYSEX              (0xF0)
1  EXTENDED_ANALOG_READ     (0x66)
2  QUERY                    (0x00)
3  pin                      (0-127)
4  END_SYSEX                (0xF7)

// In order to report analog values for pin number > 15 and/or values > 14 bits
0  START_SYSEX              (0xF0)
1  EXTENDED_ANALOG_READ     (0x66)
2  RESPONSE                 (0x01)
3  pin                      (0-127)
4  bits 0-6                 [optional] (least significant byte - not sent if value is 0)
5  bits 7-13                [optional] (most significant byte - not sent if all bits are the same as bit 6)
... additional bytes may be sent if more bits are needed (could limit this to 32 bits)
N  END_SYSEX                (0xF7)
```

The bytes containing the value won't be sent if the value is 0 or if the MSB bytes have only the same bit as the higher bit on the previous byte. This way, both signed and unsigned values can be reconstructed without having to send all the bits of the original value. When receiving, the host needs to fill all the left-most bits not received with the value from the most significant bit received.

Here's an example in JS:

```
for (var i = 0; i < numberBuffer.length; i++) {
  value = value | ((numberBuffer[i] & 0x7F) << (i * 7));
}
if (numberBuffer[numberBuffer.length - 1] & 0b01000000) {
  // Last higher bit is 1, so fill the remaining bits with 1's
  value = value | (-1 << (7 * numberBuffer.length));
}
```

I left the `// analogInputsToReport[ANALOG_INPUTS_BYTE_INDEX(analogPin)] &=  ~ (1 << ANALOG_INPUTS_BIT_INDEX(analogPin));` comments for reference as I prefer the bit set functions, but either way is working fine, let me know your preference.

### DIFF
--- a/src/AnalogInputFirmata.cpp
+++ b/src/AnalogInputFirmata.cpp
@@ -30,12 +30,12 @@ void reportAnalogInputCallback(byte analogPin, int value)
 AnalogInputFirmata::AnalogInputFirmata()
 {
   AnalogInputFirmataInstance = this;
-  analogInputsToReport = 0;
+  reset();
   Firmata.attach(REPORT_ANALOG, reportAnalogInputCallback);
 }
 
 // -----------------------------------------------------------------------------
-/* sets bits in a bit array (int) to toggle the reporting of the analogIns
+/* sets bits in a byte array to toggle the reporting of the analog pins
  */
 //void FirmataClass::setAnalogPinReporting(byte pin, byte state) {
 //}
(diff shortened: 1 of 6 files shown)


pgrawehr:
Support for Analog pins > 15 has now been added in ConfigurableFirmata 3.1. I believe this is obsolete.

Issue #81, opened 2018-05-28
setup call to systemResetCallback causes spurious analog reporting behavior


ethanjli:
# Problem
Currently, in the example sketch (and in sketches generated by Firmata Builder), `systemResetCallback()` is called at the end of `setup()` (as in [line 377 of the example ConfigurableFirmata.ino](https://github.com/firmata/ConfigurableFirmata/blob/2.10.0/examples/ConfigurableFirmata/ConfigurableFirmata.ino#L377) and in [line 316 of firmata-builder's builder.js](https://github.com/firmata/firmata-builder/blob/v0.6.1/lib/builder.js#L316)). However, this call does not trigger the other work done in the private `FirmataClass::systemReset()` method, most notably adjustment of the private `FirmataClass::resetting` member and its `FirmataClass::isResetting()` getter.

The problematic consequence is that, when `setup()` calls `systemResetCallback()` directly, it will cause `AnalogInputFirmata::reportAnalog(...)` to spuriously enter the `if (!Firmata.isResetting())` block and make a `Firmata.sendAnalog(analogPin, analogRead(analogPin))` call. I believe this is the same issue which was fixed in [issue 185 of firmata/arduino](https://github.com/firmata/arduino/issues/185). I first identified this issue when troubleshooting the same problem which was described there, except that I am using a ConfigurableFirmata sketch from Firmata Builder rather than StandardFirmata (see below).

# Proposed Solution
I believe that this issue can be solved by replacing the `systemResetCallback()` call at the end of `setup()` in `ConfigurableFirmata.ino` with either:
1. a direct call to `FirmataClass::systemReset()` (which would require making that private method public but seems more elegant to me), or
2. a call to `FirmataClass::parse(...)` with the appropriate `SYSTEM_RESET` as the argument, which has the same effect as just calling `FirmataClass::systemReset()` directly, at least when this call is made at the end of `setup()`.

I've confirmed with basic testing (see below) that option 2 seems to solve the problematic consequence described above, and it does not require making the private `FirmataClass::systemReset()` method public, simply a one-line change to `examples/ConfigurableFirmata/ConfigurableFirmata.ino` and a one-line change to [firmata/firmata-builder](https://github.com/firmata/firmata-builder)'s `lib/builder.js`. Would this be a reasonable fix?

# Reproducing the Problem
## Testing Conditions
[...]

PR #82, opened 2018-05-29
Change example sketches to avoid spurious analog reports on setup.

ethanjli:
As discussed in #81, this prevents spurious analog pin reporting behavior at the end of setup. I was able to run two basic tests of the example sketch on an Arduino Uno:

- I ran the blink script discussed in the initial report in #81 and confirmed successful execution of the script.
- I ran a blink program implemented in Snap4Arduino (running on a ChromeOS environment) and confirmed successful connection of the Arduino board and execution of the program.


soundanalogous:
I will get back to the PR later this weekend.

soundanalogous:
Also tested with johnny-five and Breakout.js without issue.

PR #94, opened 2020-04-18
Return only as many bytes as the hardware returned

pgrawehr:
This small patch fixes https://github.com/firmata/arduino/issues/444. Instead of returning a buffeer with fewer (or no) valid data than expected, we're only returning as many bytes as there actually were. 
Most clients should not be affected by this, because if the hardware is properly working, this should not happen. However it enables the possibility to use ReadByte() commands to poll for valid I2C addresses, making the equivalent of I2Cdetect work over firmata.

### DIFF
--- a/src/I2CFirmata.h
+++ b/src/I2CFirmata.h
@@ -111,6 +111,7 @@ void I2CFirmata::readAndReportData(byte address, int theRegister, byte numBytes,
     Firmata.sendString("I2C: Too many bytes received");
   } else if (numBytes > Wire.available()) {
     Firmata.sendString("I2C: Too few bytes received");
+    numBytes = Wire.available();
   }
 
   i2cRxData[0] = address;


soundanalogous:
Thanks!

soundanalogous:
Do you mind also making the same change in all the StandardFirmata example variants [here](https://github.com/firmata/arduino/tree/master/examples)?

pgrawehr:
@soundanalogous I haven't used any of those, and since they report a smaller version number than this one does, I thought they were obsolete. 
For my understanding: What is the advantage of StandardFirmata over ConfigurableFirmata? Or: Why do they exist in parallel?

soundanalogous:
StandardFirmata requires much less memory than ConfigurableFirmata with the exact same feature set enabled. If I could figure out a way to reduce that I'd merge them back together and use the modular approach for both.

pgrawehr:
@soundanalogous : Well, a bit of testing doesn't really confirm that for me. For StandardFirmata.ino, the Compiler tells me it uses 1065Bytes of RAM. For ConfigurableFirmata, I get 1165Bytes when I disable OneWire and Serial, which seem not to be part of StandardFirmata (so the feature set includes Digital in/out, Analog in/out, i2c, scheduler and reporting). This is an acceptable pay-off for the larger flexibility and better architecture IMHO. The ConfigurableFirmata OneWire driver wastes a lot of memory by allocating a struct for each possible pin.

soundanalogous:
It is nearly a 10% increase in RAM. However you don't need to include Scheduler so that saves a few bytes.

PR #96, opened 2020-05-04
Remove unused member

pgrawehr:
... that eats a lot of RAM.

### DIFF
--- a/src/OneWireFirmata.cpp
+++ b/src/OneWireFirmata.cpp
@@ -17,6 +17,14 @@
 #include "OneWireFirmata.h"
 #include "Encoder7Bit.h"
 
+OneWireFirmata::OneWireFirmata()
+{
+  for (int i = 0; i < TOTAL_PINS; i++) {
+    pinOneWire[i].device = NULL;
+    pinOneWire[i].power = false;
+  }
+}
+
 boolean OneWireFirmata::handlePinMode(byte pin, int mode)
 {
   if (IS_PIN_DIGITAL(pin) && mode == PIN_MODE_ONEWIRE) {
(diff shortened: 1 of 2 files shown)


pgrawehr:
@soundanalogous Are you accepting PR's on this project still? I'd have quite a bunch of suggestions/fixes that could make it into the library. (i.e. some more memory optimizations, SPI support, etc).

soundanalogous:
Which boards have you verified this change on?

soundanalogous:
I will consider pull requests if they have been well tested against a variety of architectures that are compatible with Firmata. I want to avoid fixes that have only been tested on one or two boards.

pgrawehr:
I have only tested the build on a nano so far, but this change is a pure code change that cannot really be hardware dependent. I removed a member of a struct (`ow_device_info::addr`) that was only ever assigned to. The value was never read.

soundanalogous:
Took a long look at this today, including the OneWire library this code depends on. Seems safe to remove that member. I'm not sure why it was ever adde in the first place. My only assumption is that an one point there may have been an intent to store the address value that gets assigned within the [search function](https://github.com/firmata/ConfigurableFirmata/blob/master/src/OneWireFirmata.cpp#L70), but that part was never added.

soundanalogous:
In general, I've been trying to step away from this project, so until someone takes over, expect long delays in code reviews unless a change is very trivial.

pgrawehr:
@soundanalogous I've been quite intensively working with firmata recently, especially writting the client library that will make it into https://github.com/dotnet/iot (See https://github.com/dotnet/iot/pull/1039, currently still in progress because of open dependencies and some low-level design discussions). I expect that this will be quite widely used, as part of a package officially supported by Microsofts dotnet team. 

Since there are a few proposed extensions that would really bring the project forward (i.e I've already got an SPI implementation), it would be a pitty to let this project starve. 

I haven't pushed to much to this repo yet (not least due to the fact that you seemed otherwise busy), but it would be an honour for me to take this over and bring it forward. Of course, I'd be extra-careful not to break stuff.

soundanalogous:
@pgrawehr please see this thread regarding getting involved: https://github.com/firmata/arduino/issues/449.

PR #108, opened 2021-07-04
Clean up Support for 32Bit Arduino Due

pgrawehr:


src/ConfigurableFirmata.cpp
```diff
@@ -489,7 +489,7 @@ void FirmataClass::sendStringf(const FlashString* flashString, int sizeOfArgs, .
 {
 	// The parameter "sizeOfArgs" is currently unused.
 	// 16 bit board?
-#if UINT_MAX <= UINT16_MAX
+#ifdef ARDUINO_ARCH_AVR
```
soundanalogous, line 492: Are you sure you want to limit this just to AVR architectures?

pgrawehr: No, I actually want to limit it to all architectures with only a "small" amount of RAM, but I didn't really find a good way of detecting that. There doesn't seem to be a way that for instance tells whether a CPU is 8 bit. I may need to add something like that to boards.h.

PR #117, opened 2021-07-24
RP2040 support - cast to PinStatus and include stdarg.h

mattjlewis:
Tested with Arduino Mbed OS RP2040 v2.3.1

### DIFF
--- a/BoardSupport.md
+++ b/BoardSupport.md
@@ -1,0 +1,77 @@
+# Notes on Specific Boards
+
+## RP2040 / Raspberry Pi Pico
+
+First of all add support for RP2040 boards in the Arduino IDE by going to Boards Manager within the Tools menu.
+Enter "RP2040" in the search box, select "Arduino Mbed OS RP2040 Boards" and install the latest version (2.3.1 at the time of writing).
+
+ConfigurableFirmata v2.10.1 requires a minor modification to the `Boards.h` file in the `src/utility` folder for the Pico to work with the Firmata libraries.
+On macOS by default this file can be found here: `~/Documents/Arduino/libraries/ConfigurableFirmata/src/utility/Boards.h`.
+
+Add the following prior to the `#else` instruction towards the end of this file to enable support for the Raspberry Pi Pico:
+
+```
+// Raspberry Pi Pico
+// https://datasheets.raspberrypi.org/pico/Pico-R3-A4-Pinout.pdf
+#elif defined(TARGET_RP2040) || defined(TARGET_RASPBERRY_PI_PICO)
+
+#include <stdarg.h>
+
+static inline void attachInterrupt(pin_size_t interruptNumber, voidFuncPtr callback, int mode)
+{
+   attachInterrupt(interruptNumber, callback, (PinStatus) mode);
+}
+
+#define TOTAL_ANALOG_PINS       4
+#define TOTAL_PINS              30
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) < 23) || (p) == LED_BUILTIN)
+#define IS_PIN_ANALOG(p)        ((p) >= 26 && (p) < 26 + TOTAL_ANALOG_PINS)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) != LED_BUILTIN)
+// From the data sheet I2C-0 defaults to GP 4 (SDA) & 5 (SCL) (physical pins 6 & 7)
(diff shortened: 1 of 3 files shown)


src/Frequency.cpp
```diff
@@ -107,7 +107,11 @@ boolean Frequency::handleSysex(byte command, byte argc, byte* argv)
 			  }
 			  if (internalMode >= 0)
 			  {
+#if defined (TARGET_RP2040) || defined (TARGET_RASPBERRY_PI_PICO)
+				  attachInterrupt(digitalPinToInterrupt(pin), FrequencyIsr, (PinStatus) internalMode);
```
pgrawehr, line 111: I think it would be better if this was also handled in boards.h, i.e. with a wrapper function. ifdefs in code files should be avoided.

src/utility/Boards.h
```diff
@@ -729,6 +729,32 @@ void analogWrite(uint8_t channel, uint32_t value, uint32_t valueMax = 255);
...
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) < 23) || (p) == LED_BUILTIN)
+#define IS_PIN_ANALOG(p)        ((p) >= 26 && (p) < 26 + TOTAL_ANALOG_PINS)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) != LED_BUILTIN)
+// From the data sheet I2C-0 defaults to GP 4 (SDA) & 5 (SCL) (physical pins 6 & 7)
+// However, v2.3.1 of mbed_rp2040 defines WIRE_HOWMANY to 1 and uses the non-default GPs 6 & 7:
+//#define WIRE_HOWMANY	(1)
+//#define PIN_WIRE_SDA            (6u)
+//#define PIN_WIRE_SCL            (7u)
+#define IS_PIN_I2C(p)           ((p) == PIN_WIRE_SDA || (p) == PIN_WIRE_SCL)
```
pgrawehr, line 748: With this setting, which pins are now used for I2C and SPI? Doesn't this need some special configuration command (I read that the RP2040 has a very flexible pinout)?

pgrawehr (changes requested):
Some minor comments

src/utility/Boards.h
mattjlewis: GPIOs 6 and 7 are used for I2C by the Mbed_rp2040 by default - that's what I tested with. I will check if they can be overridden by redefining in utility/Boards.h.

src/Frequency.cpp
mattjlewis: Fixed in the latest commit.

src/utility/Boards.h
mattjlewis: From my investigations, the I2C pins can be set by editing `./variants/RASPBERRY_PI_PICO/pins_arduino.h`. On macOS this is in `~/Library/Arduino15/packages/arduino/hardware/mbed_rp2040/2.3.1`. Yes, the Pico is very flexible regarding which pins are used for I2C. This Boards.h definition assumes on one I2C bus (I2C-1). I can investigate if it is possible to use both I2C buses.

pgrawehr: I would say the later is optional (although it would of course be nice if one could use more than one I2C bus). 

But we should document that somewhere. Maybe you start a "BoardSupport.md" file in the root directory describing the particulars of the RP2040. Documentation about how to use Firmata with which boards is apparently missing.

mattjlewis: Added notes on Pico support in a new BoardsSupport.md file.

BoardSupport.md
```diff
@@ -0,0 +1,77 @@
...
+
+## RP2040 / Raspberry Pi Pico
+
+First of all add support for RP2040 boards in the Arduino IDE by going to Boards Manager within the Tools menu.
+Enter "RP2040" in the search box, select "Arduino Mbed OS RP2040 Boards" and install the latest version (2.3.1 at the time of writing).
+
+ConfigurableFirmata v2.10.1 requires a minor modification to the `Boards.h` file in the `src/utility` folder for the Pico to work with the Firmata libraries.
+On macOS by default this file can be found here: `~/Documents/Arduino/libraries/ConfigurableFirmata/src/utility/Boards.h`.
+
+Add the following prior to the `#else` instruction towards the end of this file to enable support for the Raspberry Pi Pico:
```
pgrawehr, line 11: I'm a bit confused now. After we merge this PR, this information is obsolete, right?

mattjlewis: Yes, simply added in case anyone reads this and would like to use the current release on a Pico. This should be removed once the next version is released, I'm not sure on the timelines - days, months or years.

pgrawehr: Ok, got it. Neither do I, actually. I haven't looked into how to generate a new release at all.

pgrawehr (approved):

PR #118, opened 2021-07-27
Extend BoardSupport.md

pgrawehr:


BoardSupport.md
```diff
@@ -37,13 +37,26 @@ Official pinout diagram is here: https://content.arduino.cc/assets/Pinout-Due_la
...
+| Number of analog inputs | 4 |
+| Flash Memory | 2Mb |
+| RAM | 264kb |
+| PWM capable pins | 16 |
+| Built-in LED | Yes, pin 25 |
+| I2C, Bus 0 | Pin 6 SDA, Pin 7 SCL |
+| I2C, Bus 1 | Not currently supported by ConfigurableFirmata |
+| SPI, Bus 0 | Pin 16 MISO, Pin 17 CS, Pin 18 SCK, Pin 19 MOSI |
+| SPI, Bus 1 | Not currently supported by ConfigurableFirmata |
+
```
pgrawehr, line 59: @mattjlewis Can you confirm the above is correct for the current implementation? I have created similar tables for a several boards now.

mattjlewis: According to https://datasheets.raspberrypi.org/pico/Pico-R3-A4-Pinout.pdf and Google.
By "Logical pin numbers" I assume you are referring to "Physical pin numbers".
Number of analog inputs: 3 (the board itself has 4 - 3 are exposed on physical pins 31, 32 & 34, one is used for the internal temperature sensor hence not exposed). Note they are 12-bit resolution.
PWM capable pins - PWM is possible on all 26 GPIOs, can have 16 PWM channels at any one time.
Not sure about the SPI pins. Looking at pins_arduino.h, SPI is MISO: GP4 (physical 6), MOSI: GP3 (physical 5), SCK: GP2 (physical 4), CS: GP5 (physical 7).

mattjlewis: Also the Mbed OS tooling defaults Serial to TX: GP0 (physical 1), RX: GP1 (physical 2).

pgrawehr: No, I meant "logical pin numbers" (the GPXX numbers). You wrote that I2C-0 is on GP6 and GP7 (which is actually weird, because these pins are only for I2C-1, according to the pinout).

mattjlewis: Apologies, I should have spotted that. I have no idea what the mbed folks were thinking ignoring default values and using I2C-1. The `Wire` and `Wire1` attributes within the mbed toolchain do not have to correlate with the physical I2C bus numbers, `Wire1` can be I2C-0.

PR #128, opened 2022-03-01
The SPI configuration failed to configure the CS pin

pgrawehr:
And fix SPI buffer size
And allow 7-bit encoding for transfers (documentation update is prepared)


pgrawehr:
Related documentation update: https://github.com/firmata/protocol/pull/136

PR #144, opened 2023-04-06
fix: setFirmwareNameAndVersion before begin

awong1900:
When init call begin(115200), it call printFirmwareVersion(). But if not call setFirmwareNameAndVersion before begin, it will not report to host. So the "ready" connect will fail on J5 or Firmata.js.

Read printFirmwareVersion function:
```
    if (firmwareVersionMajor != 0 && FirmataStream != nullptr) { // make sure that the name has been set before reporting
```

### DIFF
--- a/examples/ConfigurableFirmata/ConfigurableFirmata.ino
+++ b/examples/ConfigurableFirmata/ConfigurableFirmata.ino
@@ -213,9 +213,9 @@ void initFirmata()
 
 void setup()
 {
+	initFirmata();
 	initTransport();
 	Firmata.sendString(F("Booting device. Stand by..."));
-	initFirmata();
 
 	Firmata.parse(SYSTEM_RESET);
 


pgrawehr:
Ok, I understand the problem you're facing, however I think this is only hiding a bug in the host. Even with the change, you're relying on the board resetting during connect. This is usually the case for Arduino UNO boards, but not for more modern hardware. The host should instead send a REPORT_FIRMWARE request after connecting.

awong1900:
> The host should instead send a REPORT_FIRMWARE request after connecting.
Yes, you are right. The Firmata.js simply detect that no protocol version has been received, the protocol version request will be send. However, if the protocol version is detected and the firmware version is not detected, it will not send firmware version request, which will result in an unsuccessful connection.

I have do a test. Both use standard ConfigurableFirmata.ino on UNO and ESP32.
On UNO,  the board send protocol version, not send firmware version. So host do not re request, connect fail.
On ESP32, the board not send protocol version and firware version, So host do re request, connect success.

Issue #136, opened 2023-02-01
Things that should already be fixed in ConfigurableFirmata


jcvillegasfernandez:
### Bugs reported long ago and not fixed yet:
Fix the compatibility problem of FirmataScheduler and OnewireFirmata with devices other than Arduino Uno, like ESP8266, ESP32 and probably others due to different sizes of pointer, Int and long.

### New bug detected:
A bug (hard to detect) when the size of the SYSEX command to send is 64 (buffer size), the last byte is lost, works fine with any other value, solution at the end.

### Modify file FirmataScheduler.cpp in this way

inside this function
boolean FirmataScheduler::handleSysex(byte command, byte argc, byte* argv)
...
case DELAY_FIRMATA_TASK:
          {
            if (argc == 6) {
              argv++;
              Encoder7BitClass::readBinary(4, argv, argv); //decode inplace
     **// long is different on ESP8266 (or ESP32) and arduino**
              _delayTask((long)argv[0] | (long)argv[1]<<8 | (long)argv[2]<<16 | (long)argv[3]<<24);
            }
            break;
          }
        case SCHEDULE_FIRMATA_TASK:
          {
            if (argc == 7) { //one byte taskid, 5 bytes to encode 4 bytes of long
              Encoder7BitClass::readBinary(4, argv + 2, argv + 2); //decode inplace
    **// long is different on ESP8266 (or ESP32) and arduino**
              _schedule(argv[1], (long)argv[2] | (long)argv[3]<<8 | (long)argv[4]<<16 | (long)argv[5]<<24);
            }
            break;
          }
....
void FirmataScheduler::reportTask(byte id, firmata_task* task, boolean error)
...
if (task) {
        encoder.startBinaryWrite();
    **// because long and int are different on ESP8266 (or ESP32) and arduino**
            // write time_ms
        encoder.writeBinary(task->time_ms & 0xFF); 
        encoder.writeBinary((task->time_ms >> 8) & 0xFF);  
        encoder.writeBinary((task->time_ms >> 16) & 0xFF); 
        encoder.writeBinary((task->time_ms >> 24) & 0xFF); 
           // write len
        encoder.writeBinary(task->len & 0xFF); 
        encoder.writeBinary((task->len >> 8) & 0xFF);
            // write pos
        encoder.writeBinary(task->pos & 0xFF); 
        encoder.writeBinary((task->pos >> 8) & 0xFF); 
            // write messages
[...]

Issue #89, opened 2019-06-12
OneWireFirmata does not work on esp8266 and solution


jcvillegasfernandez:
Trying my Lazarus client over wifi I realized that the module OneWireFirmata does not work on esp8266, exception issue (some thing about a bad alignment), but OneWire library worked.

After a lot of researching I think I have discovered the wrong code:

  if (subcommand & ONEWIRE_READ_REQUEST_BIT) {
     if (numBytes < 4) break;
// this is the bad part
   numReadBytes = *((int*)argv);
   argv += 2;
   correlationId = *((int*)argv);
   argv += 2;
// this is the right one
  numReadBytes = argv[0] | ((int)argv[1] << 8);
  correlationId = argv[2] | ((int)argv[3] << 8);
  argv += 4;

and you have to change these lines too:
  if (subcommand & ONEWIRE_DELAY_REQUEST_BIT) {
         if (numBytes < 4) break;  
// bad code
   Firmata.delayTask(*((long*)argv));
// this is the right one
   Firmata.delayTask(argv[0] | ((long)argv[1] << 8) | ((long)argv[2] << 16) | ((long)argv[3] << 24));

kormizz:
Thank you, it works, great work!!!

jcvillegasfernandez:
I only try to use configurablefirmata for my Lazarus client and if I find something wrong when I test my  client I try to fix it but I am not an arduino programmer.

lawern:
Thank you, it works for me too!

echavet:
https://github.com/firmata/ConfigurableFirmata/issues/136#issuecomment-1518248362

pgrawehr:
This should be fixed with #136. Please reopen if not.

echavet:
I know this is a closed issue but...

This is a bit surprising, given that all the platforms (esp8266, ardiono uno or zero for me) in question are little endian and the "long" data type is consistently 32 bits in size. Despite this, users have reported that the proposed code change resolves the issue!

For me, this problem arises when working with Arduino Uno, but not with Arduino Zero. Interestingly, the occurrence of the error seems to be contingent upon the inclusion of FirmataScheduler, as defined by the configuration typedef constant:

```cpp
#ifdef ENABLE_BASIC_SCHEDULER
```

Looking for an explanation, I discovered that the implementation of the delayTaskCallback function is specific to FirmataScheduler. It calls `Firmata.attachDelayTask(delayTaskCallback);` for its software delay implementation:

[...]

PR #148, opened 2023-04-23
#136 : sizeOf Int depends on the board, use of int16_t OneWireFirmata…

echavet:
…::handleSysex

Several users reports memory pb with the OneWire handleSysex method.  I've tested this change on arduino Zero. No more freeze.

### DIFF
--- a/src/OneWireFirmata.cpp
+++ b/src/OneWireFirmata.cpp
@@ -116,11 +116,11 @@ boolean OneWireFirmata::handleSysex(byte command, byte argc, byte* argv)
 
                 if (subcommand & ONEWIRE_READ_REQUEST_BIT) {
                   if (numBytes < 4) break;
-                  numReadBytes = *((int*)argv);
-                  argv += 2;
-                  correlationId = *((int*)argv);
-                  argv += 2;
-                  numBytes -= 4;
+                  numReadBytes = *((int16_t*)argv);
+                  argv += sizeof(int16_t);
+                  correlationId = *((int16_t*)argv);
+                  argv += sizeof(int16_t);
+                  numBytes -= 2 * sizeof(int16_t);
                 }
 
                 if (subcommand & ONEWIRE_DELAY_REQUEST_BIT) {


pgrawehr:
Looking good, thanks. I do not know why the build is currently failing. This is certainly not related to the changes.

pgrawehr:
The problem is the arduino-esp32 core, which breaks the build in v 2.0.8. Should hopefully be fixed soonish.

pgrawehr (approved):

pgrawehr:
Note: Arduino-esp32 core 2.0.8 is not supported. Use 2.0.9 or later.

echavet:
There remains an issue, potentially related to the one resolved [here](https://github.com/firmata/ConfigurableFirmata/issues/89#issue-455098232), for which [jcvillegasfernandez](https://github.com/jcvillegasfernandez) has proposed a solution. The suggested fix was to replace this line:

```cpp
Firmata.delayTask(((long)argv));
```

with this:

```cpp
Firmata.delayTask(argv[0] | ((long)argv[1] << 8) | ((long)argv[2] << 16) | ((long)argv[3] << 24));
```

This is a bit surprising, given that all the platforms (esp8266, ardiono uno or zero for me) in question are little endian and the "long" data type is consistently 32 bits in size. Despite this, users have reported that the proposed code change resolves the issue!

For me, this problem arises when working with Arduino Uno, but not with Arduino Zero. Interestingly, the occurrence of the error seems to be contingent upon the inclusion of FirmataScheduler, as defined by the configuration typedef constant:

```cpp
#ifdef ENABLE_BASIC_SCHEDULER
```

Looking for an explanation, I discovered that the implementation of the delayTaskCallback function is specific to FirmataScheduler. It calls `Firmata.attachDelayTask(delayTaskCallback);` for its software delay implementation:

```cpp
void FirmataScheduler::delayTask(long delay_ms)
{
  if (running) {
    long now = millis();
    running->time_ms += delay_ms;
    if (running->time_ms < now) { // If delay time already passed, schedule to 'now'.
      running->time_ms = now;
    }
  }
}
```

I am no longer able to test this, but I believe that without `ENABLE_BASIC_SCHEDULER`, there is NO implementation for `delayTaskCallback`. Therefore, the delay operation is solely dependent on the runtime execution of the `delayTask` method:

```cpp
if (delayTaskCallback) {
  (*delayTaskCallback)(delay);
}
```

Now let's consider the comparison between these two scenarios:

1. Three left shifts and three long casts: `argv[0] | ((long)argv[1] << 8) | ((long)argv[2] << 16) | ((long)argv[3] << 24)`
2. Versus a single long cast: `Firmata.delayTask(*((long*)argv));`

Given that I'm using the Johnny-Five library, and its implementation of the DS18B20 thermometer uses `board.io.sendOneWireDelay(pin, 1);` to allow for temperature conversion, it's plausible that this delay is primarily necessary for slower platforms.

My hypothesis is that there would be no discernible difference with a longer delay period when `delayTaskCallback` is null because there is no delay instruction.

```cpp
/**
 * Call the delayTask callback function when using FirmataScheduler. Must first attach a callback
 * using attachDelayTask.
 * @see FirmataScheduler
 * @param delay The amount of time to delay in milliseconds.
 */
void FirmataClass::delayTask(long delayMs) {
  if (delayTaskCallback) {
    (*delayTaskCallback)(delayMs);
  } 
  // adding a default implementation seems to solve the issue
   else {
    delay(delayMs);
  }
}
```

---

PR #157, opened 2023-09-04
SPI: fix bug in csPin data type

logicog:
The data type of the csPin member of spi_device_config was wrong. The field will be set to -1 if no pin is set or manual cs-pin control is used. Then the comparisons with -1 in the code will not work as expected, see https://stackoverflow.com/questions/36770376/comparing-unsigned-char-to-a-negative-number

So we use "char" instead of "byte" and the code behaves as expected.


pgrawehr (approved):

pgrawehr:
Thank you for this fix!

Issue #151, opened 2023-06-28
Issue handling zero-length SYSEX message


mchesser:
The code doesn't handle zero length SYSEX messages correctly. Sending the message `[START_SYSEX, END_SYSEX]`/`[0xf0, 0xf7]`, causes `processSysexMessage` to be called ([here](https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L326)) with `sysexBytesRead == 0`. 

I think this causes two issues:

* First, this will causes the previous SYSEX command to be executed, because `storedInputData[0]` is not valid:

https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L216

* Second, an underflow occurs when computing `sysexBytesRead - 1`. If the previous SYSEX command was as `STRING_DATA` command then the code here [ConfigurableFirmata.cpp:229](https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L229-L237) could overflow the `storedInputData` buffer (though only without `LARGE_MEM_DEVICE` defined, since the index is a byte).

In practice, I guess clients shouldn't generate these messages but I thought it was worth reporting.

pgrawehr:
Thanks for the report. 

I agree this is a minor issue, but should still be looked at.

PR #159, opened 2023-11-07
Handle zero-length SYSEX messages

mchesser:
Simple fix for #151. Check that the `storedInputData` buffer is not empty before reading the command ID from the buffer.

### DIFF
--- a/src/ConfigurableFirmata.cpp
+++ b/src/ConfigurableFirmata.cpp
@@ -213,6 +213,10 @@ int FirmataClass::available(void)
  */
 void FirmataClass::processSysexMessage(void)
 {
+  if (sysexBytesRead == 0) {
+    return;
+  }
+
   switch (storedInputData[0]) { //first byte in buffer is command
     case REPORT_FIRMWARE:
       printFirmwareVersion();


pgrawehr:
Thanks, this is looking good. Just curious: Is there any reasonable use case where this happens? Or did you find this just based on a bug in the client library?

echavet:
De mémoire et en ayant jeté un oeil rapide, sauf erreur ce cas n'est pas possible. La fonction n'est appelée que lorqu'un message sysex complet est déjà dans le tampon.

mchesser:
> Thanks, this is looking good. Just curious: Is there any reasonable use case where this happens? Or did you find this just based on a bug in the client library?

The actual issue was originally found via some fuzzing experiments I was doing. But I was envisioning an API like `SendSysex(data)` that automatically adds `SYSEX_START/SYSEX_END` to the data, that could be accidentally misused by providing an empty buffer to the function. I also thought it might be possible for a byte to be dropped because line corruption. Since this issue is somewhat difficult to debug, I thought it would be nice to fix. 

I had quick look through the other implementations, and the .NET IoT implementation does contain a similar `bytes_read` check when parsing SYSEX messages [FirmataDevice.cs:359-364](https://github.com/dotnet/iot/blob/c754c2cc5bd9e8fcd3cbe258e304e2546885f0cc/src/devices/Arduino/FirmataDevice.cs#L359-L364) 

> De mémoire et en ayant jeté un oeil rapide, sauf erreur ce cas n'est pas possible. La fonction n'est appelée que lorqu'un message sysex complet est déjà dans le tampon. Mais bon au cas où on l'appellerait en dehors du contexte actuel, la modif est bonne.

If the bytes `[START_SYSEX, END_SYSEX]` are sent, for the first byte the following code is executed: [ConfigurableFirmata.cpp:407-410](https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L407-L410), for the second byte the code here is executed: [ConfigurableFirmata.cpp:322-326](https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L322-L326). Neither of those two blocks of code ever write to `storedInputData` so `storedInputData` is effectively empty (the prefix/suffix bytes are never written to the buffer). Another option is to move the check to [Line 326](https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L326) instead. 


_(J'ai dû recourir à Google Translate pour répondre, désolé si c'est difficile à comprendre), Si les octets `[START_SYSEX, END_SYSEX]` sont envoyés : pour le premier octet le code suivant est exécuté : [ConfigurableFirmata.cpp:407-410](https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L407-L410), pour le deuxième octet, le code ici est exécuté : [ConfigurableFirmata.cpp:322-326](https://github.com/firmata/ConfigurableFirmata/blob/c759e5a0375bdbb89efb0c7efe640dccd3305f65/src/ConfigurableFirmata.cpp#L322-L326). Aucun de ces deux blocs de code n'écrit jamais dans `storeInputData`, donc `storeInputData` est effectivement vide (les octets de préfixe/suffixe ne sont jamais écrits dans le tampon). Une autre option consiste à déplacer le chèque vers [Line 326](https://github.com/firmata/ConfigurableFirmata

pgrawehr:
There appears to be a compatibility issue with the ESP IDK 3.0. I need to look into that first.

pgrawehr:
@mchesser Can you please rebase your PR? That should now fix the build.

mchesser:
Rebased.

pgrawehr:
Thanks again. Merged.

PR #162, opened 2024-02-25
Update ConfigurableFirmata.ino 

PizzaProgram:
Adding option to: 
 - Start all output pins at "HIGH state" 

to the example file.

### DIFF
--- a/examples/ConfigurableFirmata/ConfigurableFirmata.ino
+++ b/examples/ConfigurableFirmata/ConfigurableFirmata.ino
@@ -137,6 +137,8 @@ void systemResetCallback()
 		else if (IS_PIN_DIGITAL(i)) 
 		{
 			Firmata.setPinMode(i, PIN_MODE_OUTPUT);
+// Uncomment following line, if you want to start the board at HIGH output state on powering up:
+//			digitalWrite(pin, HIGH);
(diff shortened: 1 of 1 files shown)


examples/ConfigurableFirmata/ConfigurableFirmata.ino
```diff
@@ -137,6 +137,8 @@ void systemResetCallback()
 		else if (IS_PIN_DIGITAL(i)) 
 		{
 			Firmata.setPinMode(i, PIN_MODE_OUTPUT);
+// Uncomment following line, if you want to start the board at HIGH output state on powering up:
+//			digitalWrite(pin, HIGH);
```
pgrawehr, line 141: You should also call `Firmata.setPinState()` to make sure the internal pin state cache is updated.

PizzaProgram: Good idea! Thanks for the tip! 
Also I see I have made 2 big mistakes:  
- Forgot the `Firmata.` part, 
- wrote `pin,` instead of `i,`

PizzaProgram:
Done
https://github.com/firmata/ConfigurableFirmata/pull/163

PR #163, opened 2024-02-26
Update ConfigurableFirmata.ino

PizzaProgram:
Adding option to:
- Start all output pins at "HIGH state"

to the example file.

### DIFF
--- a/examples/ConfigurableFirmata/ConfigurableFirmata.ino
+++ b/examples/ConfigurableFirmata/ConfigurableFirmata.ino
@@ -136,7 +136,10 @@ void systemResetCallback()
 		} 
 		else if (IS_PIN_DIGITAL(i)) 
 		{
-			Firmata.setPinMode(i, PIN_MODE_OUTPUT);
+			Firmata.setPinMode  (i, PIN_MODE_OUTPUT);
+// Uncomment following 2 lines, if you want to start the board at HIGH output state on powering up:
+//			Firmata.digitalWrite(i, HIGH);
+//			Firmata.setPinState (i, HIGH);
 		}
 	}
 #endif


pgrawehr (approved):

pgrawehr:
Thanks for this addition

PizzaProgram:
I need more help on this, can not solve it alone. 🥺 
Also I need to apologize for commiting too soon.

I've just tried this code I've commited in real life, and 
 - it has no effect !

I do not really understand why, and no matter what I try it makes things just worse.
_(If I remember well, it worked with standardfirmata + Arduino Nano board combination.)_

That means, if I apply this patch with ConfigurableFirmata + Raspberry Pi PICO board,
and plug in the board:
1. first it gives a little ground to all pins, (the LED of the relay board starts to glow with 25% strength) 
2. than a few seconds later, after the "version-led-blink": 
3. it turns all the output pins LOW (and by that, turns on all the relays, what i would like to avoid).

So there are 2 problems:
 - it does not execute immediately on powering the board (probably because the pin state is floating)
 - and it is somehow resetting to LOW state.

I've tried to copy the set-high code to the setup part too, but than the board goes into an infinite boot loop. 
_(Blinking shorts and logs non-stop.)_
```c
void setAllHIGH() {
PinStatus p = HIGH;

	for (byte i = 0; i < TOTAL_PINS; i++) {
		if (IS_PIN_DIGITAL(i)) {
		//if (i > 2) {p = LOW;} else {p = HIGH;} ;  // uncomment this for testing (distunguish 1+2. pin and all the others)
			digitalWrite       (i, p);
			Firmata.setPinMode (i, PIN_MODE_OUTPUT);
			digitalWrite       (i, p);
			Firmata.setPinState(i, p);
			digitalWrite       (i, p);
		}
	}
}

```

pgrawehr:
First question: What board are you using? 

The Arduino AVR-based boards boot with all pins in INPUT mode, so yes, they're floating. You should be able to add external pull-up resistors to make the input high at startup (and thus turn off the relays). 

I can't say why it would then switch to low when the Startup code runs, I'll need to try.

PizzaProgram:
> What board are you using?

[RPi Pico](https://www.raspberrypi.com/documentation/microcontrollers/raspberry-pi-pico.html) (see my prev. msg)

> The Arduino AVR-based boards boot with all pins in INPUT mode, so yes, they're floating.

_Yes, I know. It seems also this board does._
I've hoped to be able to turn HIGH fast enough during boot.
_(Soldering 60+ resistors & wires is not really a dream job.)_

> I can't say why it would then switch to low when the Startup code runs, I'll need to try.

Thank You! 😄

pgrawehr:
I just tried with an Arduino Nano, and this code works for me (all outputs high after boot phase):

```
		digitalWrite(i, HIGH);
		Firmata.setPinState (i, HIGH);
```

PizzaProgram:
> I just tried with an Arduino Nano, and this code works for me (all outputs high after boot phase)

So basically you just uncommented the 2 lines at my original Commit, and it works "out of the box" ?
- Or did you placed those 2 lines somewhere elsewhere?

pgrawehr:
No, I just uncommented your lines (except that the first line should not have the `Firmata.` prefix) - that causes a compile error.

PizzaProgram:
After investigating the problem from all sides, I've realized:
 - It is **not safe** to work with "inverted pins" from software side!
 - This must be solved on the hardware side!

to prevent any accidents.

_There are 2 possibilities:_

## 1. Use PNP / NPN inverter boards:
https://duckduckgo.com/?t=ftsa&q=PNP+NPN+opto+board&iax=images&ia=images

## 2. Use "High-triggered" Relay boards.

https://duckduckgo.com/?q=high+trigger+relay+board&t=ftsa&iar=images&iax=images&ia=images

... or those kind, which have jumpers to where someone can choose between Low trigger / High trigger
kép

See those yellow jumpers on the picture.

So this commit should be reverted fully.
 _(There are also 3 other problems with my code.)_

pgrawehr:
Now I'm curious: Why is it not safe? Can you explain that?

PizzaProgram:
> Now I'm curious: Why is it not safe? Can you explain that?  
  
< OFF >  
For example:  
 - the software is freezing,
 - electric disturbance,
 - the separate 5V power of the arduino / RPi board dies (and the relay board keeps being powered).
  
Connecting heavy metal roller-shutters on those relays which can basically cut a head off, or "over-driving a heater" because it did not turn off...  
Many things can go wrong even in a simple house, not to mention any industrial application.  
_(And believe me: Firmata is being used already for such tasks too!)_

pgrawehr:
Yikes... I do prefer if my head stays exactly where it is...

PR #174, opened 2024-11-22
Fix `#if` in AnalogOutputFirmata.h

Noorquacker:
The `void internalReset();` private member function of the AnalogOutputFirmata class is intended to only be defined if compiling on ESP32.

No other code in this library uses `#if ESP32`, they all use `#ifdef ESP32`. When changing this to use `#ifdef`, this successfully compiles in Arduino IDE 2.3.2. Previously, when using `#if`, this wasn't compiling and the compiler complained about there being no member function defined for the class.

Hopefully this works in other situations and doesn't break anything else.

### DIFF
--- a/src/AnalogOutputFirmata.h
+++ b/src/AnalogOutputFirmata.h
@@ -30,7 +30,7 @@ class AnalogOutputFirmata: public FirmataFeature
     void analogWriteInternal(byte pin, uint32_t value);
   private:
       void setupPwmPin(byte pin);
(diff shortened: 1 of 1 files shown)


pgrawehr:
Thanks a lot for the suggested fix. Coincidentaly, a similar patch for this issue already had a PR open at #170 which is now merged.

pgrawehr:
The other PR is now merged.

PR #184, opened 2025-04-09
Prevent out-of-bounds access of servo array

### LINKED ISSUE #176: Servo on Raspberry Pi Pico
When `#define ENABLE_SERVO` is enabled on the Raspberry Pi Pico, the board fails to start.
The issue occurs because `MAX_SERVOS` is defined as 8, while `TOTAL_PINS` is 30. This causes an out-of-range access in the `ServoFirmata::reset()` function.

I'm not using servos in my project, so I have disabled this functionality. As a result, I cannot confirm the exact behavior in detail.

maxgerhardt:
Attempt to fix #176. 

For the RPi Pico, `MAX_SERVOS` = 8 (using the Arduino-Pico core's Servo library) and `TOTAL_PINS = 30`.

The code just assumes that on every pin that *can* be a servo (`IS_PIN_SERVO()`), it can also use that pin as an index into the `servos` array (8 elements long) and call detach on it. This is wrong. In iteration pin = 8, it will grab `servos[8]` (an out of bounds read of that array), get back a not-null-pointer and call detach on that object, which hardfaults the Pico's processor.

This simple fix generically aborts execution if the index into the servos array would be out of bounds. For me, it then doesn't crash anymore.

See also discussion in https://github.com/earlephilhower/arduino-pico/discussions/2897.

### DIFF
--- a/src/ServoFirmata.h
+++ b/src/ServoFirmata.h
@@ -130,6 +130,10 @@ void ServoFirmata::attach(byte pin, int minPulse, int maxPulse)
 
 void ServoFirmata::detach(byte pin)
 {
+  // Servo not available for this pin?
+  if (PIN_TO_SERVO(pin) >= sizeof(servos) / sizeof(servos[0])) {
+    return;
+  }
   Servo *servo = servos[PIN_TO_SERVO(pin)];
   if (servo) {
     if (servo->attached())


maxgerhardt:
Mhm, while this prevents the crash, the logic doesn't seem to be 100% right.. That also prevents a detach of the servo pin for all pins >= 8.

Of course, you could have a Servo object on a pin higher than that (e.g., 16), and it could e.g. use the first available servo object in the `servos` array. Something here is wrong in the base logic. The 
```cpp
#define PIN_TO_SERVO(p)         (p)
```
does not account for that.

PizzaProgram:
IMHO adding 1 more line would would make it even safer at the beginning:
 `if (! IS_PIN_SERVO(pin)) {return;}`

```c
void ServoFirmata::detach(byte pin)
{
  // Servo not available for this pin?
  if (! IS_PIN_SERVO(pin)) {return;}
  if (PIN_TO_SERVO(pin) >= sizeof(servos) / sizeof(servos[0])) {
    return;
  }
  Servo *servo = servos[PIN_TO_SERVO(pin)];
  Servo *servo = servos[PIN_TO_SERVO(pin)];
```

pgrawehr:
Ups, didn't intend to merge so quickly... But shouldn't hurt anyway. Please create a new PR for any further changes.

